PR #47, opened 2020-07-31
[R4R] Validator count to be enforced in the smart contracts

guagualvcha:
Currently, the number of BSC validators is not limited in the contracts.  The max intended number of BSC validators is 41.


contracts/BSCValidatorSet.sol
```diff
@@ -395,6 +396,9 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
   /*********************** Internal Functions **************************/
 
   function checkValidatorSet(Validator[] memory validatorSet) private pure returns(bool, string memory) {
+    if (validatorSet.length > MAX_NUM_OF_VALIDATORS){
+      return (false, "number of validators exceed limit");
```
yutianwu, line 400: can we change the comment to:

`the number of validators exceeds the limit`

guagualvcha: sure

PR #54, opened 2020-07-31
R4R: add upper limit into claimReward

HaoyangLiu:


contracts/TokenHub.sol
```diff
@@ -105,6 +105,9 @@ contract TokenHub is ITokenHub, System, IParamSubscriber, IApplication, ISystemR
 
   function claimRewards(address payable to, uint256 amount) onlyInit onlyRelayerIncentivize external override returns(uint256) {
     uint256 actualAmount = amount < address(this).balance ? amount : address(this).balance;
+    if (actualAmount > 1e18) {
```
guagualvcha, line 108: `1e18 ` to be a const then?

PR #71, opened 2020-08-15
R4R: add name method to bep2e interface

HaoyangLiu:


guagualvcha:
do not forget about change the doc

PR #120, opened 2021-12-28
[R4R] implement BEP-127

SolidityGo:
# BEP-127: Temporary Maintenance Mode for Validators

- [BEP-127: Temporary Maintenance Mode for Validators](#bep-99-temporary-maintenance-mode-for-validators)
    - [1. Summary](#1-summary)
    - [2. Abstract](#2-abstract)
    - [3. Status](#3-status)
    - [4. Motivation](#4-motivation)
    - [5. Specification](#5-specification)
    - [6. License](#6-license)

## 1. Summary
This BEP introduces the **Temporary Maintenance** mode for validators on the Binance Smart Chain.

## 2. Abstract
**Temporary Maintenance** is supposed to last one or a few hours.  The validator seat will be temporarily dropped from the block producing rotation during the maintenance. Since long-time offline maintenance is not encouraged, the validator will still be slashed if the maintenance lasts too long. To lower the impact from poorly-operating validators who forget to claim its maintenance, they will be forced to enter **Temporary Maintenance** mode too.

## 3. Status
Draft.

## 4. Motivation
Due to the design of **Parlia** consensus, the absence of the validator, even if it is due to scheduled maintenance, will cause instability and capacity loss of BSC due to the extra waiting time and chain reorganization for other validators. Introducing **Temporary Maintenance** mode will stabilize the blocking rotation and maintain the capacity of BSC.

### Current Slash Mechanisms
The [slash contract](https://bscscan.com/address/0x0000000000000000000000000000000000001001) will record the missed blocking metrics of each validator. 
- Once the metrics are above the predefined threshold, the blocking reward for validator will not be relayed to BC for distribution but shared with other better validators.
- If the metrics remain above another higher level of threshold, the validator will be dropped from the rotation, and put as “in jail”. This will be propagated back to Binance Chain, where a predefined amount of BNB would be slashed from the self-delegated BNB of the validator. 

#### Proactive Maintenance
Validator can claim itself to enter scheduled maintenance by sending a transaction signed by the consensus key. The validator can claim itself to exit maintenance by sending another transaction. The slash cleaning work will be done within the exit transaction:

**SlashCount =  (MaintenanceEndHeight - MaintenanceStartHeight)/len(currentValidatorSet)/Scale**

**Scale** is a governable parameter, the initial setting is 2, usually it should be larger than 1.  If **SlashCount** is larger than a predefined value, the validator will still be slashed. The validator can get more time to bring the node back before being slashed if it claims itself to enter maintenance. Validator is encouraged to claim  itself to exit maintenance even if it will be put in jail, since it can send the unjail transaction earlier.

#### Passive Maintenance
Once the number of missed blocks is above a predefined threshold, the validator will enter maintenance automatically. The validator still gets a chance to catch up and claim itself online again. 

### Limit
1. The number of maintained validators has an upper limit at the same time. Once exceeded, the following proactive/passive claim request will be rejected.
2. A validator can only enter maintenance once per day.

### ValidatorSet Change
The validators election will repeat every 24 hours, the maintenance will end immediately once election happens.

### Impact to Validator Maintainers
Validator maintainers can claim the validator to enter maintenance mode once the node runs into unexpected issues. Validator maintainers need to send transactions to exit maintenance mode even if the validator is enforced into maintenance, otherwise, the validator may be put in jail.

## 6. License

The content is licensed under [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


contracts/BSCValidatorSet.sol
```diff
@@ -73,6 +73,26 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     Validator[] validatorSet;
   }
 
+  // BEP-99 Temporary Maintenance
+  uint256 public constant EXIT_MAINTENANCE_REWARD_RATIO_SCALE = 100;
```
guagualvcha, line 77: Please place field together, while struct with struct.

contracts/BSCValidatorSet.sol
```diff
@@ -73,6 +73,26 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+  uint256 public constant EXIT_MAINTENANCE_REWARD_RATIO = 20;
+  uint256 public constant SAFE_MIN_GAS = 2300;
```
guagualvcha, line 79: `constant` should at top of the file

contracts/BSCValidatorSet.sol
```diff
@@ -479,6 +558,64 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+      maintainingValidatorSet[index] = maintainingValidatorSet[maintainingValidatorSet.length - 1];
+      maintainInfoMap[maintainingValidatorSet[index].consensusAddress].maintainingIndex = index;
+    }
+    maintainingValidatorSet.pop();
+
+    // TODO add event
+  }
+
+  function _safeTransferBNB(address to, uint256 value) internal {
+    (bool success, ) = to.call{ gas: SAFE_MIN_GAS, value: value }("");
```
guagualvcha, line 615: I suggest we simply use `to.transfer(value)`, the relayer can not be contract, we have limit on relayerhub.

contracts/BSCValidatorSet.sol
```diff
@@ -73,6 +73,26 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     Validator[] validatorSet;
   }
 
+  // BEP-99 Temporary Maintenance
+  uint256 public constant EXIT_MAINTENANCE_REWARD_RATIO_SCALE = 100;
+  uint256 public constant EXIT_MAINTENANCE_REWARD_RATIO = 20;
+  uint256 public constant SAFE_MIN_GAS = 2300;
+
+  uint256 public MaxNumOfMaintaining;
+  uint256 public MaxMaintainingTime;
```
guagualvcha, line 82: `MaxMaintainingTime` or `MaxMaintainingBlocks`

contracts/BSCValidatorSet.sol
```diff
@@ -73,6 +73,26 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+
+  struct MaintainInfo {
+    // TODO optimize struct
+    bool isMaintaining;
+    uint256 maintainStartAt;
+    uint256 maintainingIndex;   // the index of the element in `maintainingValidatorSet`.
+    uint256 exitMaintenanceReward;
```
guagualvcha, line 89: change to :
```
  struct MaintainInfo {
    uint256 startAt; // Must be larger than 0
    uint256 index;   // the index of the element in `maintainingValidatorSet`.
    uint256 exitReward;
}
```
if `startAt`==0, it means not in maintain.

contracts/BSCValidatorSet.sol
```diff
@@ -73,6 +73,26 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+  }
+  // key is the `consensusAddress` of `Validator`,
+  mapping(address => MaintainInfo) public maintainInfoMap;
```
guagualvcha, line 92: `maintainInfoMap` ==> `maintainingValidatorSetMap`

contracts/BSCValidatorSet.sol
```diff
@@ -400,6 +433,52 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     // averageDistribute*rest may less than income, but it is ok, the dust income will go to system reward eventually.
   }
 
+  /*********************** For Temporary Maintenance **************************/
+  function canEnterMaintenance(address validator, uint256 index) public view returns (bool can, uint256 exitMaintenanceReward) {
```
guagualvcha, line 437: `address validator` is useless.

contracts/BSCValidatorSet.sol
```diff
@@ -400,6 +433,52 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+    Validator memory validatorInfo = currentValidatorSet[index];
+
+    if (
+      (MaxNumOfMaintaining == 0 || MaxMaintainingTime == 0)         // - 1. check if not start
+      || maintainingValidatorSet.length >= MaxNumOfMaintaining      // - 2. check if exceeded upper limit
+      || validatorInfo.jailed                                       // - 3. check if jailed
+      || maintainInfoMap[validator].isMaintaining                   // - 4. check if maintaining
+      || maintainInfoMap[validator].maintainStartAt > 0             // - 5. check if has Maintained
+      || currentValidatorSet.length <= 1                            // - 6. check num of remaining currentValidators
```
guagualvcha, line 446: damn. I find a bug in previous implement, `currentValidatorSet.length <= 1` need exclude the validator that is in jail and in maintaining .

guagualvcha: You need change the implementation  of `getValidators` and use `len(getValidators()) <=1` as condition here.

contracts/BSCValidatorSet.sol
```diff
@@ -347,7 +376,11 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     }
     // the actually index
     index = index - 1;
-    uint256 income = currentValidatorSet[index].incoming;
+
+    // deduct the exitMaintenanceReward
+    (bool can, uint256 exitMaintenanceReward) = canEnterMaintenance(validator, index);
+
+    uint256 income = currentValidatorSet[index].incoming.sub(exitMaintenanceReward);
```
guagualvcha, line 383: if can && exitMaintenanceReward> 0, then do `sub`.

contracts/BSCValidatorSet.sol
```diff
@@ -400,6 +433,52 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+
+  function enterMaintenance() external {
+    uint256 index = currentValidatorSetMap[msg.sender];
+    require(index > 0, "only current validators can enter maintenance");
+
+    // the actually index
+    index = index - 1;
+    (bool can, uint256 exitMaintenanceReward) = canEnterMaintenance(msg.sender, index);
+    require(can, "can not enter Temporary Maintenance");
+    _enterMaintenance(msg.sender, index, exitMaintenanceReward);
```
guagualvcha, line 463: we can not remove it from currentValidatorSet. If we do, when the validatorSet rotation happen at UTC 00, we wont get the message about the income of this validator.

contracts/BSCValidatorSet.sol
```diff
@@ -400,6 +433,52 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+  }
+
+  function exitMaintenance(address validator) external {
+    MaintainInfo memory maintainInfo = maintainInfoMap[validator];
+    require(maintainInfo.isMaintaining && maintainInfo.maintainStartAt > 0, "validator is not maintaining");
+
+    // disable reentry while _exitMaintenance
+    require(msg.sender == tx.origin, "no proxy is allowed");
```
guagualvcha, line 471: we can delete this line because we should require the reward searcher to be a relayer, a relayer must be a EOA

contracts/BSCValidatorSet.sol
```diff
@@ -73,6 +73,26 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     Validator[] validatorSet;
   }
 
+  // BEP-99 Temporary Maintenance
+  uint256 public constant EXIT_MAINTENANCE_REWARD_RATIO_SCALE = 100;
+  uint256 public constant EXIT_MAINTENANCE_REWARD_RATIO = 20;
```
yutianwu, line 78: should we add this into governance

contracts/BSCValidatorSet.sol
```diff
@@ -479,6 +558,64 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+
+    // TODO add event
+  }
+
+  function _exitMaintenance(address validator, address payable rewardSearcher) private {
+    // step 1: modify status of the validator
+    MaintainInfo storage maintainInfo = maintainInfoMap[validator];
+    maintainInfo.isMaintaining = false;
+    if (maintainInfo.exitMaintenanceReward > 0) {
+      _safeTransferBNB(rewardSearcher, maintainInfo.exitMaintenanceReward);
```
guagualvcha, line 593: if the sender is validator, we can not simply send BNB to him back.

contracts/BSCValidatorSet.sol
```diff
@@ -366,6 +399,10 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
       }
     }
     // averageDistribute*rest may less than income, but it is ok, the dust income will go to system reward eventually.
+
+    if (can) {
```
yutianwu, line 403: can we refactor this `can`, it's far away from where it's defined

contracts/BSCValidatorSet.sol
```diff
@@ -73,6 +73,26 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     Validator[] validatorSet;
   }
 
+  // BEP-99 Temporary Maintenance
+  uint256 public constant EXIT_MAINTENANCE_REWARD_RATIO_SCALE = 100;
+  uint256 public constant EXIT_MAINTENANCE_REWARD_RATIO = 20;
+  uint256 public constant SAFE_MIN_GAS = 2300;
+
+  uint256 public MaxNumOfMaintaining;
```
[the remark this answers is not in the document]
yutianwu: where is `MaxNumOfMaintaining` initialized

contracts/BSCValidatorSet.sol
```diff
@@ -208,6 +228,15 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
     }
 
+    // step 0: force all maintaining validators to exit `Temporary Maintenance`
+    address validator;
+    for (uint i = 0; i < maintainingValidatorSet.length; i++) {
+      validator = maintainingValidatorSet[i].consensusAddress;
+      _exitMaintenance(validator, payable(BURN_ADDRESS));
+      delete maintainInfoMap[validator];
+    }
+    delete maintainingValidatorSet;
```
guagualvcha, line 238: Are you sure we can simply delete an slice in this way?

guagualvcha:
After I walk through the code, my main concerns are:
1. Fund losing forever. If a validator claim it self to be maintained, we should let him pay a percentage of income to system reward contract, but keep most of the income for him. The fund is lost for current implement.
2. Slash Record did not increase when the validator been forced on line.

contracts/BSCValidatorSet.sol
SolidityGo:
```
`startAt`==0 means has not maintained  during this 24 hours period of time.
if `isMaintaining` removed,  there's no way to know if it's maintaining

1. before enterMaintenance:  `startAt`==0,   `isMaintaining` == false
2. Maintaining: `startAt` != 0,   `isMaintaining` == true
3. after exitMaintenance:   `startAt` != 0,   `isMaintaining` == false



```

contracts/BSCValidatorSet.sol
SolidityGo: maintainInfoMap  includes  3 status info (`not maintained` / `maintaining` / `maintained` ),  `maintainingValidatorSetMap` may not be accurate

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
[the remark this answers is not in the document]
SolidityGo: > the first character should be lower.

fixed

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
SolidityGo: removed code

contracts/BSCValidatorSet.sol
SolidityGo: yes

SolidityGo:
> After I walk through the code, my main concerns are:
> 
> 1. Fund losing forever. If a validator claim it self to be maintained, we should let him pay a percentage of income to system reward contract, but keep most of the income for him. The fund is lost for current implement.
> 2. Slash Record did not increase when the validator been forced on line.

contracts/BSCValidatorSet.sol
SolidityGo: removed code

contracts/BSCValidatorSet.sol
SolidityGo: removed code

contracts/BSCValidatorSet.sol
SolidityGo: income of validators will recorded in `maintainingValidatorSet` while validators enter `Temporary Maintenance`

contracts/BSCValidatorSet.sol
SolidityGo: removed code

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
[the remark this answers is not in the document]
SolidityGo: > where is `MaxNumOfMaintaining` initialized

in updateParam

contracts/BSCValidatorSet.sol
```diff
@@ -201,13 +220,21 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
   }
 
   function updateValidatorSet(Validator[] memory validatorSet) internal returns (uint32) {
-    // do verify.
-    (bool valid, string memory errMsg) = checkValidatorSet(validatorSet);
-    if (!valid) {
-      emit failReasonWithStr(errMsg);
-      return ERROR_FAIL_CHECK_VALIDATORS;
+    {
```
forcodedancing, line 223: Sorry, not get it - why here needs extra "{" ?

contracts/BSCValidatorSet.sol
```diff
@@ -341,63 +367,49 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+  /*********************** For Temporary Maintenance **************************/
+  function canEnterMaintenance(uint256 index) public view returns (bool) {
+    Validator memory validatorInfo = currentValidatorSet[index];
+    address validator = validatorInfo.consensusAddress;
+
+    if (
+      (maxNumOfMaintaining == 0 || maintainSlashScale == 0)         // - 1. check if not start
+      || maintainingValidatorSet.length >= maxNumOfMaintaining      // - 2. check if exceeded upper limit
+      || validatorInfo.jailed                                       // - 3. check if jailed
+      || maintainInfoMap[validator].isMaintaining                   // - 4. check if maintaining
```
forcodedancing, line 389: When maintainInfoMap's data is populated? Is it possible maintainInfoMap[validator] == null here?

contracts/BSCValidatorSet.sol
```diff
@@ -97,6 +114,8 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
   event unexpectedPackage(uint8 channelId, bytes msgBytes);
   event paramChange(string key, bytes value);
   event feeBurned(uint256 amount);
+  event validatorMaintain(address indexed validator);
```
guagualvcha, line 117: `validatorMaintain` ==> `validatorEnterMaintenance`

contracts/BSCValidatorSet.sol
```diff
@@ -341,63 +367,49 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
-    if (index <= 0) {
-      return;
+  function felony(address validator)external onlySlash override {
+    _felony(validator);
+  }
+
+  /*********************** For Temporary Maintenance **************************/
+  function canEnterMaintenance(uint256 index) public view returns (bool) {
+    Validator memory validatorInfo = currentValidatorSet[index];
+    address validator = validatorInfo.consensusAddress;
```
guagualvcha, line 383: We can pass `validator` as a parameter of `canEnterMaintenance`

guagualvcha: For example, if index is `100`, it will return true.

SolidityGo: https://github.com/binance-chain/bsc-genesis-contract/pull/120#discussion_r776134581

contracts/BSCValidatorSet.sol
```diff
@@ -341,63 +367,49 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
-    index = index - 1;
-    uint256 income = currentValidatorSet[index].incoming;
-    currentValidatorSet[index].incoming = 0;
-    uint256 rest = currentValidatorSet.length - 1;
-    emit validatorMisdemeanor(validator,income);
-    if (rest==0) {
-      // should not happen, but still protect
-      return;
+    uint256 validatorIndex = _misdemeanor(validator);
+    if (canEnterMaintenance(validatorIndex)) {
```
guagualvcha, line 371: Need check `validatorIndex`

contracts/BSCValidatorSet.sol
```diff
@@ -479,7 +501,178 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+    // step 1: modify status of the validator
+    MaintainInfo storage maintainInfo = maintainInfoMap[validator];
+    maintainInfo.isMaintaining = false;
+
+    // step 2: add the validator to currentValidatorSet
+    uint256 index = maintainInfo.index;
+    currentValidatorSet.push(maintainingValidatorSet[index]);
+    currentValidatorSetMap[validator] = currentValidatorSet.length;
+
+    // step 3: remove the validator from maintainingValidatorSet
```
guagualvcha, line 641: Shall we remove it? otherwise, the validator can enter maintainence multi times within one day.

contracts/BSCValidatorSet.sol
SolidityGo: // scope to avoid stack too deep errors

contracts/BSCValidatorSet.sol
```diff
@@ -479,7 +501,178 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+  function _forceMaintainingValidatorsExit(Validator[] memory validatorSet) private returns (Validator[] memory unjailedValidatorSet){
+    uint256 numOfFelony = 0;
+    address validator;
+    bool isFelony;
+
+    // 1. validators exit maintenance
+    for (uint i = 0; i < maintainingValidatorSet.length; i++) {
+      validator = maintainingValidatorSet[i].consensusAddress;
+      // exit maintenance
+      isFelony = _exitMaintenance(validator);
```
guagualvcha, line 582: No need to `_exitMaintenance` if it is not `isMaintaining`

contracts/BSCValidatorSet.sol
SolidityGo: If the validator entered maintenance, `|| maintainInfoMap[validator].startBlockNumber > 0`, this check condition will not pass, then the validator can only enter maintenance once within one day

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
SolidityGo: each validator in `maintainingValidatorSet ` must be `isMaintaining`

contracts/BSCValidatorSet.sol
```diff
@@ -479,7 +501,178 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+    if (index != maintainingValidatorSet.length - 1) {
+      maintainingValidatorSet[index] = maintainingValidatorSet[maintainingValidatorSet.length - 1];
+      maintainInfoMap[maintainingValidatorSet[index].consensusAddress].index = index;
+    }
+    maintainingValidatorSet.pop();
+    emit validatorExitMaintenance(validator);
+
+    // step 4: slash
+    uint256 slashCount =
+      block.number.sub(maintainInfo.startBlockNumber).div(currentValidatorSet.length).div(maintainSlashScale);
```
yutianwu, line 653: the `maintainSlashScale` may be zero

contracts/BSCValidatorSet.sol
```diff
@@ -479,7 +501,178 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     return v1.consensusAddress == v2.consensusAddress && v1.feeAddress == v2.feeAddress && v1.BBCFeeAddress == v2.BBCFeeAddress && v1.votingPower == v2.votingPower;
   }
 
+  function _misdemeanor(address validator) private returns (uint256) {
+    uint256 index = currentValidatorSetMap[validator];
```
yutianwu, line 505: if `index` == 0 or `index` == 1, it will both return 0

contracts/BSCValidatorSet.sol
```diff
@@ -479,7 +501,178 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+      }
+    }
+    // averageDistribute*rest may less than income, but it is ok, the dust income will go to system reward eventually.
+
+    return index;
+  }
+
+  function _felony(address validator) private {
+    uint256 index = currentValidatorSetMap[validator];
+    if (index <= 0) {
```
yutianwu, line 537: if index == 0 or index == 1, it will both return 0

contracts/BSCValidatorSet.sol
SolidityGo: If maintainInfoMap's data is not populated, it defaults to the initial value in solidity

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
SolidityGo: fixed,  it should not happen, but still protect

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
SolidityGo: > Please pass `validator` as a parameter rather than `index`. For current implementation, it will return invalid result if the index is quite large.

 Passing `validator` as a parameter  will cause a lot to change, and the `pragma experimental ABIEncoderV2;
` needs to be imported.
So I've now added the index check to fix this problem

contracts/BSCValidatorSet.template
```diff
@@ -32,7 +34,7 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
   uint256 public constant EXPIRE_TIME_SECOND_GAP = 1000;
   uint256 public constant MAX_NUM_OF_VALIDATORS = 41;
 
-  bytes public constant INIT_VALIDATORSET_BYTES = hex"{{initValidatorSetBytes}}";
+  bytes public constant INIT_VALIDATORSET_BYTES = hex"f84580f842f840949fb29aac15b9a4b7f17c3385939b007540f4d791949fb29aac15b9a4b7f17c3385939b007540f4d791949fb29aac15b9a4b7f17c3385939b007540f4d79164";
```
yutianwu, line 37: why change this in template

contracts/BSCValidatorSet.template
```diff
@@ -52,10 +54,18 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
 
   uint256 public constant BURN_RATIO_SCALE = 10000;
   address public constant BURN_ADDRESS = 0x000000000000000000000000000000000000dEaD;
-  uint256 public constant INIT_BURN_RATIO = {{initBurnRatio}};
+  uint256 public constant INIT_BURN_RATIO = 0;
```
yutianwu, line 57: and this

contracts/BSCValidatorSet.sol
```diff
@@ -56,6 +58,14 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
   uint256 public burnRatio;
   bool public burnRatioInitialized;
 
+  // BEP-99 Temporary Maintenance
```
guagualvcha, line 61: `BEP-99` ==> `BEP-127`, let's search all BEP-99, and replace them

contracts/BSCValidatorSet.template
SolidityGo: I changed it by mistake

contracts/BSCValidatorSet.template
SolidityGo: I changed it by mistake, fixed

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
```diff
@@ -341,63 +372,54 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
 
   /*********************** For slash **************************/
   function misdemeanor(address validator)external onlySlash override{
-    uint256 index = currentValidatorSetMap[validator];
-    if (index <= 0) {
-      return;
+    uint256 validatorIndex = _misdemeanor(validator);
```
guagualvcha, line 382: We need change the logic of slash contract, to ensure no one can jump `misdemeanor`

contracts/BSCValidatorSet.sol
```diff
@@ -341,63 +372,54 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
-      }
-    }
-    // averageDistribute*rest may less than income, but it is ok, the dust income will go to system reward eventually.
+
+    require(canEnterMaintenance(index), "can not enter Temporary Maintenance");
+    _enterMaintenance(msg.sender, index);
+  }
+
+  function exitMaintenance() external {
+    require(maintainInfoMap[msg.sender].isMaintaining, "not in maintenance");
```
guagualvcha, line 428: We need more strong verification here, check whether it exist in `maintainingValidatorSet`

contracts/BSCValidatorSet.sol
```diff
@@ -413,6 +435,16 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
       require(newBurnRatio <= BURN_RATIO_SCALE, "the burnRatio must be no greater than 10000");
       burnRatio = newBurnRatio;
       burnRatioInitialized = true;
+    } else if (Memory.compareStrings(key, "maxNumOfMaintaining")) {
+      require(value.length == 32, "length of maxNumOfMaintaining mismatch");
+      uint256 newMaxNumOfMaintaining = BytesToTypes.bytesToUint256(32, value);
```
guagualvcha, line 447: Can we have an initial value for `maxNumOfMaintaining` and `maintainSlashScale`, just like the bruno burn?

guagualvcha: Also please add some testcase to change the value of these.

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
SolidityGo: fixed

contracts/BSCValidatorSet.sol
SolidityGo: Ok,   initial values added

contracts/BSCValidatorSet.sol
```diff
@@ -479,7 +546,190 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+    }
+
+    uint256 slashCount =
+      block.number.sub(maintainInfo.startBlockNumber).div(currentValidatorSet.length).div(maintainSlashScale);
+
+    (uint256 misdemeanorThreshold, uint256 felonyThreshold) = ISlashIndicator(SLASH_CONTRACT_ADDR).getSlashThresholds();
+    isFelony = false;
+    if (slashCount >= felonyThreshold) {
+      _felony(validator);
+      ICrossChain(CROSS_CHAIN_CONTRACT_ADDR).sendSynPackage(SLASH_CHANNELID, encodeSlashPackage(validator), 0);
```
guagualvcha, line 716: Validator contract do not have the permission to send `SLASH_CHANNELID` package, now only slash contract is allowed to do so, you may check the `crosschain` contract.

SolidityGo: fixed

contracts/SlashIndicator.sol
```diff
@@ -80,6 +80,7 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
 
   /*********************** External func ********************************/
   function slash(address validator) external onlyCoinbase onlyInit oncePerBlock onlyZeroGasPrice{
+    require(IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).isCurrentValidator(validator), "not current validator");
```
guagualvcha, line 83: Can we do not use require but simply return, since `slash` is called by consensus engine,  we don't want consensus engine fail in any case.

SolidityGo: fixed

guagualvcha (approved):

contracts/BSCValidatorSet.template
```diff
@@ -201,13 +226,21 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
   }
 
   function updateValidatorSet(Validator[] memory validatorSet) internal returns (uint32) {
-    // do verify.
-    (bool valid, string memory errMsg) = checkValidatorSet(validatorSet);
-    if (!valid) {
-      emit failReasonWithStr(errMsg);
-      return ERROR_FAIL_CHECK_VALIDATORS;
+    {
```
KeefeL, line 229: No need change here, the code style should keep consistent.

contracts/BSCValidatorSet.template
```diff
@@ -422,6 +476,16 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
       burnRatioInitialized = true;
+    } else if (Memory.compareStrings(key, "maxNumOfMaintaining")) {
+      require(value.length == 32, "length of maxNumOfMaintaining mismatch");
+      uint256 newMaxNumOfMaintaining = BytesToTypes.bytesToUint256(32, value);
+      require(newMaxNumOfMaintaining < MAX_NUM_OF_VALIDATORS, "the maxNumOfMaintaining must be less than MAX_NUM_OF_VALIDATORS");
+      maxNumOfMaintaining = newMaxNumOfMaintaining;
+    } else if (Memory.compareStrings(key, "maintainSlashScale")) {
+      require(value.length == 32, "length of maintainSlashScale mismatch");
+      uint256 newMaintainSlashScale = BytesToTypes.bytesToUint256(32, value);
+      require(newMaintainSlashScale > 0, "the maintainSlashScale must be greater than 0");
```
KeefeL, line 487: I think the maintainSlashScale should no less than 1?

KeefeL (approved):
LGTM

yutianwu (approved):

PR #145, opened 2022-05-31
Fast finality

realuncle:


contracts/SystemReward.sol
```diff
@@ -54,4 +57,20 @@ contract SystemReward is System, ISystemReward{
...
     return operators[addr];
   }
+
+  function updateParam(string calldata key, bytes calldata value) onlyGov external override {
+    if (Memory.compareStrings(key, "updateOperator")) {
+      bytes memory valueLocal = value;
+      require(valueLocal.length == 20, "length of value for updateOperator should be 20");
+      address operatorAddr;
+      assembly {
+        operatorAddr := mload(add(valueLocal, 20))
```
AndyJiangPro, line 67: Consider using `abi.decode`. I think the code is more readable using `abi.decode/encode`.
1. No need to validate its length
2. Get rid of assembly

PR #189, opened 2022-09-22
update: fix some audit issues

Pythonberg1997:
### Description

Add more balance and address check to make the module more robust.

### Rationale

N/A

### Example

N/A

### Changes

- Use `call` instead of `transfer` to send BNB
- Add more balance check

### DIFF
--- a/contracts/Staking.sol
+++ b/contracts/Staking.sol
@@ -398,6 +398,11 @@ contract Staking is IStaking, System, IParamSubscriber, IApplication {
       uint256 newMinDelegation = BytesToTypes.bytesToUint256(32, value);
       require(newMinDelegation > relayerFee, "the minDelegation must be greater than relayerFee");
       minDelegation = newMinDelegation;
+    } else if (Memory.compareStrings(key, "transferGas")) {
+      require(value.length == 32, "length of transferGas mismatch");
+      uint256 newTransferGas = BytesToTypes.bytesToUint256(32, value);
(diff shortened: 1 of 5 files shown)


axljoker03:
> ### Description
> 
> Add more balance and address check to make the module more robust.
> 
> ### Rationale
> 
> N/A
> 
> ### Example
> 
> N/A
> 
> ### Changes
> 
> - Use `call` instead of `transfer` to send BNB
> - Add more balance check
>

guagualvcha (approved):

contracts/Staking.sol
```diff
@@ -40,10 +40,12 @@ contract Staking is IStaking, System, IParamSubscriber, IApplication {
   uint256 public constant INIT_RELAYER_FEE = 16 * 1e15;
   uint256 public constant INIT_BSC_RELAYER_FEE = 1 * 1e16;
   uint256 public constant INIT_MIN_DELEGATION = 100 * 1e18;
+  uint256 public constant INIT_TRANSFER_GAS = 2300;
```
guagualvcha, line 43: Please pay attention here if we want to upgrade testnet to the latest code version.

axljoker03: thank you

MyTrueWallet (changes requested):
This changes is used by hackers only except some of you trying to change this codes are actually thieves and are already hacked into several people Crypto wallets.

MyTrueWallet (approved):

PR #192, opened 2022-10-07
fix: add height check for cross chain package

Pythonberg1997:
### Description
This PR tries to increase the cost of cross-chain vulnerability exploitation.

### Rationale
The hacker registered as a `relayer` and exploit the merkle proof vulnerability,  the BC block header that he use is an ancient block which obviously does not have the latest packages,  this PR will introduce the header check to prevent such situation
### Example
NA

### Changes

- Add `channelSyncedHeaderMap` to record the most recent synced header's height. New packages' height cannot be smaller than the recorded height.

### DIFF
--- a/contracts/CrossChain.sol
+++ b/contracts/CrossChain.sol
@@ -35,7 +35,7 @@ contract CrossChain is System, ICrossChain, IParamSubscriber{
   mapping(uint8 => uint64) public channelReceiveSequenceMap;
   mapping(uint8 => bool) public isRelayRewardFromSystemReward;
 
-  // to prevent forged header
+  // to prevent the utilization of ancient block header
   mapping(uint8 => uint64) public channelSyncedHeaderMap;
 
   // event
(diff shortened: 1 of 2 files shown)


cosinlink (dismissed):

cosinlink (commented):
I'm not sure if this update will cause issues with cross-chain transactions that are currently being squeezed, affecting `BSCValidatorSet` updates or `SlashIndicator`

owen-reorg (dismissed):

yutianwu (dismissed):

contracts/CrossChain.sol
```diff
@@ -35,6 +35,9 @@ contract CrossChain is System, ICrossChain, IParamSubscriber{
   mapping(uint8 => uint64) public channelReceiveSequenceMap;
   mapping(uint8 => bool) public isRelayRewardFromSystemReward;
 
+  // to prevent forged header
```
guagualvcha, line 38: `to prevent forged header` is not ture,  `to prevent the utilization  of ancient block header`

guagualvcha (dismissed):

contracts/CrossChain.sol
Pythonberg1997: Fixed

guagualvcha (approved):

brilliant-lx (approved):

PR #198, opened 2022-10-10
enhance: add whitelabel relayers

guagualvcha:
### Description
This PR will add a whitelabel relayer list, only whitelabel relayer can relay packages.

### Rationale
The hacker register as a `relayer` and exploit the mekle-proof vulnerability,  before we get further enhancement, we will close the
channel for community relayers.

### Example
NA

### Changes
NA

PR #205, opened 2022-11-15
BEP-174: Cross Chain Relayer Management

emailtovamos:
# BEP-174: Cross Chain Relayer Management

## 1. Summary

This BEP introduces a new governance proposal type to manage the set of whitelisted Relayers.  
A Relayer whitelist was introduced recently to the BSC genesis contracts ([PR-198](https://github.com/bnb-chain/bsc-genesis-contract/pull/198)). 
This BEP will improve the management of the relayer whitelist by introducing Relayer Managers, where a Relayer Manager can manage the registration of a single Relayer. 
Managers will be elected and ejected via governance.

## 2. Abstract

This BEP introduces a new governance proposal type to manage Relayer Manager public keys that will be stored in the [`RelayerHub`](https://github.com/bnb-chain/bsc-genesis-contract/blob/71dcd4c409a68a6e084645a9f1f80adecd5a4269/contracts/RelayerHub.sol) genesis contract on BSC.  

Relayer Managers will be able to manage their individual relayer public key.  Only Relayer public keys registered via Managers will be able to invoke the `RelayerHub` smart contract for cross chain transfers.  

## 3. Status

This BEP is under progress. See https://github.com/bnb-chain/bsc-genesis-contract/pull/205.

## 4. Motivation

After the [BSC Bridge exploitation](https://github.com/verichains/public-audit-reports/discussions/17) it is important we introduce some mechanism in place that enhances the security of the system. A governance based set Manager public keys who manage a single Relayer adds an extra layer of security.

### 5.1 Current Relayer Whitelist

Currently whitelisted relayers are [hardcoded](https://github.com/bnb-chain/bsc-genesis-contract/blob/a8476b2aefba0a66db86311ab44d43d7e2df24fe/contracts/System.sol#L88) in the `System` genesis smart contract.  This list needs to be updated manually by deploying a new version of the smart contract.

### 5.2 The New Whitelisting Mechanism
After the implementation of this BEP:

- A Relayer Manager role will be introduced with corresponding public keys.   
- The set of Managers will be updated via governance using the existing [param change proposal](https://github.com/bnb-chain/bnb-chain.github.io/blob/master/docs/learn/bsc-gov-workflow.md#submit-cross-chain-param-change-proposal) mechanism.
- Managers will NOT have to deposit any BNB (as opposed to the previous 100BNB deposit requirement). 
- Once selected by governance, Managers will be able to add an associated Relayer public key.
- There will be a 1:1 relationship between Manager public key to active Relayer public key.
- Only associated Relayers can send cross chain messages. 
- Managers can modify public key of their Relayer at any time. This will facilitate rotation of Relayer instances.
- The old `unregister` function will still be active to allow the existing whitelisted relayers to safely exit.

#### Relayer Manager functions

```
function updateParam(string calldata key, bytes calldata value) external override onlyInit onlyGov
```
The pre-existing function will be updated to support `addManager` and `removeManager` keys to facilitate adding and removing of managers via governance.

```
function removeManagerByHimself() external
```
Allows a manager to remove himself if required.

```
fuction addManagerByGov(address) external onlyGov
```
Adds a Manager public key to the current set of Managers. This is only called via governance. Emits an event on success.

```
function updateRelayer(address relayerToBeAdded) public onlyManager
```
This can be used by the manager to add or remove its relayer. Overwrites Relayer if there is an existing Relayer already added. Emits an event on success. The `relayerToBeAdded` parameter can be set to `0` address to enable removal of a relayer (e.g. if the private key gets compromised).

```
function whitelistInit() external 
```
This is to ensure smooth transition so that the presently existing hardcoded whitelisted relayers continue to be relayers after the hardfork. This function needs to be **manually** called once after the hardfork.

#### Relayer functions
```
function isRelayer(address relayerAddress) external override view returns (bool)
```
This will be called by the Relayer codebase, instead of registering.  Relayer can stop if this returns false

#### `updateParam` Changes

```
  function updateParam(string calldata key, bytes calldata value) external override onlyInit onlyGov{
    if (Memory.compareStrings(key,"addManager")) {
       // TODO check and parse value
       // addManagerAddress(...)
    } else if (Memory.compareStrings(key,"removeManager")) {
       // TODO check and parse value
       // removeManagerAddress(...)
    } else {
      require(false, "unknown param");
    }
    emit paramChange(key, value);
  }
```
The `updateParam` function will be modified accept two new keys
- `addManager` to add a Manager to the stored Manager set.
- `removeManager` to remove a Manager from the stored Manager set.

Also `requiredDeposit` and `dues` will be removed. So no need to account for them in the keys check.

#### Functions to be Deprecated

These Relayer functions will be deprecated:
```
function register() external payable ...
```

`unregister()` function isn't deprecated to allow existing hardcoded whitelisted relayers to have safe exit.

### 5.4 Relayer Changes

Update Relayer codebase to no longer register and deposit. Update abi? and run code generation to call new `verifyRelayer` method on `RelayerHub` smart contract and proceed to relay cross chain messages on success.

### 5.5 Existing relayer key migration

Manager keys will need to be supplied for the existing two relayers.  Manger set will be seeded with these two Manager public keys.  Relayer public keys will be registered to these Managers.

Existing Relayers will need to pull down relayer code once the `RelayerHub` SC has been upgraded.

### 5.6 CLI commands
```
## mainet
./bnbcli params  submit-bscParam-change-proposal  --key "addManager" --value "ADDRESS"   --target 0x0000000000000000000000000000000000001006 --deposit 200000000000:BNB     --voting-period 100   --side-chain-id  bsc  --title "add new super trusty relayer"  --from alice  --trust-node   --chain-id Binance-Chain-Tigris

## testnet
./tbnbcli params  submit-bscParam-change-proposal  --key "addManager" --value "ADDRESS"   --target 0x0000000000000000000000000000000000001006 --deposit 200000000000:BNB     --voting-period 100   --side-chain-id  chapel  --title "add new super trusty relayer"  --from alice  --trust-node   --chain-id Binance-Chain-Ganges
```
The commands used to create a change proposal to governance.

## 7. License

The content is licensed under [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


contracts/RelayerHub.sol
```diff
@@ -83,13 +62,90 @@ contract RelayerHub is IRelayerHub, System, IParamSubscriber{
...
+
+  function addAdminAddress(address adminToBeAdded) external onlyGov{
+    require(!relayAdminsExistMap[adminToBeAdded], "admin already exists");
+
+    relayAdminsExistMap[adminToBeAdded] = true;
+
+    emit addAdminAddress(adminToBeAdded);
+  }
+
+  function registerAdmin() external payable onlyAdmin {
```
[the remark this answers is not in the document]
emailtovamos: I decoupled the current modifier logic to registered and nonRegistered admin. Let me know if this makes sense.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,182 @@ import "./System.sol";
...
+
+    function init() external onlyNotInit {
+        requiredDeposit = INIT_REQUIRED_DEPOSIT;
+        dues = INIT_DUES;
+        alreadyInit = true;
+
+        // todo initialise the currently existing admins and their relayer keys
+
+        admins[WHITELIST_1] = admin(requiredDeposit, dues);
+        admins[WHITELIST_2] = admin(requiredDeposit, dues);
```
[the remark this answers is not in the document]
emailtovamos: This is inside the `init()` function which will be executed only during the deployment time to have a smooth transfer of logic since these are already doing relay work now. Or is there a flaw in this understanding?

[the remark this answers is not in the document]
emailtovamos: Oh I thought they were relayers before the hotfix was made after the hack i.e. they already deposited 100BNB before October?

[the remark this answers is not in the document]
bnb-twu: I'm pretty sure those hardcoded relayers had already deposited.  And if the balance of this smart contract address has enough tokens, it should be fine to allow them to unregister through their associated Manager key now.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,182 @@ import "./System.sol";
...
+    function removeRelayer() external onlyRegisteredAdmin {
+        require(adminsAndRelayers[msg.sender], "relayer doesn't exist for this admin");
+
+        relayer memory r = adminsAndRelayers[msg.sender];
+
+        delete (relayerExistsMap[r]);
+        delete (adminsAndRelayers[msg.sender]);
+
+        emit removeRelayer(r);
+    }
```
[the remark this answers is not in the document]
bnb-twu: Isn't that why we introduced `registerAdminAddRelayer`?  Are you talking about the situation where a Manager wants to change the associated Relayer key?

bnb-twu:
@emailtovamos we should be removing the existing hardcoded whitelist in `System.Sol` as well.

contracts/RelayerHub.sol
```diff
@@ -17,59 +17,38 @@ contract RelayerHub is IRelayerHub, System, IParamSubscriber{
...
-
-  modifier noProxy() {
-    require(msg.sender == tx.origin, "no proxy is allowed");
-    _;
-  }
-
-  modifier noExist() {
-    require(!relayersExistMap[msg.sender], "relayer already exist");
-    _;
-  }
```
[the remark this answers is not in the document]
emailtovamos: Is this resolved with the latest changes or do I need to change something ? :)

contracts/RelayerHub.sol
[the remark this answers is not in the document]
bnb-twu: as discussed, let's just let them overwrite their current relayer and rename `addRelayer` to `editRelayer`

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,193 @@ import "./System.sol";
...
+        managers[WHITELIST_1] = manager(requiredDeposit, dues);
+        managers[WHITELIST_2] = manager(requiredDeposit, dues);
+
+        managersRegistered[WHITELIST_1] = true;
+        managersRegistered[WHITELIST_2] = true;
+
+        relayManagersExistMap[WHITELIST_1] = true;
+        relayManagersExistMap[WHITELIST_2] = true;
+
+        managersAndRelayers[WHITELIST_1] = WHITELIST_1; // fixme current relayer
```
alexlucaci, line 76: why is the manager address used as relayer address at init?

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,193 @@ import "./System.sol";
...
-
-  function isRelayer(address sender) external override view returns (bool) {
-    return relayersExistMap[sender];
-  }
+contract RelayerHub is IRelayerHub, System, IParamSubscriber {
+    using SafeMath for uint256;
+
+    uint256 public constant INIT_REQUIRED_DEPOSIT = 1e20;
+    uint256 public constant INIT_DUES = 1e17;
+    address public constant WHITELIST_1 = 0xb005741528b86F5952469d80A8614591E3c5B632;
```
alexlucaci, line 16: I would rename the whitelist address constants to something more descriptive as it's quite unclear if what kind of addresses they are.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,193 @@ import "./System.sol";
...
+
+        address r = managersAndRelayers[msg.sender];
+
+        delete (relayerExistsMap[r]);
+        delete (managersAndRelayers[msg.sender]);
+
+        emit removeRelayerEvent(r);
+    }
+
+    function isRelayer(address relayerAddress) external override view returns (bool){
```
alexlucaci, line 197: This function is never used inside the contract. I guess there are 2 options:

- Delete it and also remove the `IRelayerHub` interface from the contract declaration (I don't know what are the implications though)
- Use it in the other functions instead of directly using the `relayerExistsMap`

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,193 @@ import "./System.sol";
...
+        managersRegistered[WHITELIST_1] = true;
+        managersRegistered[WHITELIST_2] = true;
+
+        relayManagersExistMap[WHITELIST_1] = true;
+        relayManagersExistMap[WHITELIST_2] = true;
+
+        managersAndRelayers[WHITELIST_1] = WHITELIST_1; // fixme current relayer
+        managersAndRelayers[WHITELIST_2] = WHITELIST_2; // fixme current relayer
+
+        // fixme initialise relayerExistsMap
```
alexlucaci, line 79: was this done?

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,193 @@ import "./System.sol";
...
+        registerManager();
+        addRelayer(relayer);
+    }
+
+    function removeRelayer() external onlyRegisteredManager {
+        if (managersAndRelayers[msg.sender] == address(0)) {
+            require(false, "relayer doesn't exist for this manager");
+        }
+
+        address r = managersAndRelayers[msg.sender];
```
alexlucaci, line 189: I would keep variable naming consistent, in the other function that is using the same logic the `r` variable is named `relayerAddress`

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,193 @@ import "./System.sol";
...
+    function removeManagerByGov(address payable managerToBeRemoved) internal {
+        removeManagerHelper(managerToBeRemoved);
+    }
+
+    function removeManager() external onlyRegisteredManager {
+        // here the manager removes himself
+        removeManagerHelper(msg.sender);
+    }
+
+    function removeManagerHelper(address payable managerAddress) internal {
```
alexlucaci, line 121: I would rename this to `removeManager` and use another name for the function above, something to match its purpose (manager removes himself). A couple of options could be:

- `removeManagerByHimself`
- `removeMe`
- `removeMeFromManagers`

contracts/RelayerHub.sol
guagualvcha: I will add some color here: the whitelist relayer currently is the account that is defined in the genesis files.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,189 @@ import "./System.sol";
...
+    mapping(address => address) managersAndRelayers;
+    mapping(address => bool) relayerExistsMap;
+
+    struct manager {
+        uint256 deposit;
+        uint256 dues;
+    }
+
+    modifier onlyNonRegisteredManager() {
+        require(relayManagersExistMap[msg.sender], "manager does not exist");
```
guagualvcha, line 34: `onlyNonRegisteredManager` is confusing, since it will ensure the sender is a relayer manager.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,189 @@ import "./System.sol";
...
+    event addManagerByGovEvent(address _addedManager);
+    event registerManagerEvent(address _registeredManager);
+    event addRelayerEvent(address _relayerToBeAdded);
+    event removeRelayerEvent(address _removedRelayer);
+
+
+    function init() external onlyNotInit {
+        requiredDeposit = INIT_REQUIRED_DEPOSIT;
+        dues = INIT_DUES;
+        addInitRelayer(WHITELIST_1);
```
guagualvcha, line 63: The `init` will never call again, so the modification here is useless, it means `addInitRelayer`  never done.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,189 @@ import "./System.sol";
...
+            require(false, "unknown param");
+        }
+        emit paramChange(key, value);
+    }
+
+    function removeManagerByGov(address payable managerToBeRemoved) internal {
+        removeManagerHelper(managerToBeRemoved);
+    }
+
+    function removeManager() external onlyRegisteredManager {
```
guagualvcha, line 110: If you check there `relayerHub` on mainnet, it locks 1400 BNB, it means there are 14 register relayers right now, but only two of them are whitelist relayers, we need support the rest 12 relayer to exit.

guagualvcha, line 110: Is this `onlyRegisteredManager` modifier necessary, since `removeManagerByGov` dont have this, and `removeManagerHelper` will also do duplicate check.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,189 @@ import "./System.sol";
...
+        systemPayable.transfer(a.dues);
+
+        delete (managers[managerAddress]);
+        delete (managersRegistered[managerAddress]);
+
+        // emit success event
+        emit removeManagerByGovEvent(managerAddress);
+        if (relayerAddress != address(0)) {
+            emit removeRelayerEvent(relayerAddress);
+        }
```
guagualvcha, line 136: Forget to delete `relayerExistsMap`?

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,189 @@ import "./System.sol";
...
+    }
+
+    function registerManager() internal onlyNonRegisteredManager {
+        require(msg.value == requiredDeposit, "deposit value is not exactly the same");
+        managers[msg.sender] = manager(requiredDeposit, dues);
+        managersRegistered[msg.sender] = true;
+        emit registerManagerEvent(msg.sender);
+    }
+
+    function addRelayer(address relayerToBeAdded) public onlyRegisteredManager noProxy {
```
guagualvcha, line 155: Is `noProxy` necessary here? since you already ensure that the manager can not be a contract.

guagualvcha, line 155: `addRelayer` is not a good name, since it can also update relayer.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,189 @@ import "./System.sol";
...
+        require(!relayerExistsMap[relayerToBeAdded], "relayer already exists");
+        require(!isContract(relayerToBeAdded), "contract is not allowed to be a relayer");
+
+        if(managersAndRelayers[msg.sender] != address(0)) {
+            address r = managersAndRelayers[msg.sender];
```
guagualvcha, line 160: A better practice to implement a role update:
1. No need to delete, just update it;
2. No need to emit to event(one delete, another add), just `event updateRelayer(from addres, to address)`

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,189 @@ import "./System.sol";
...
+        relayerExistsMap[relayerToBeAdded] = true;
+        emit addRelayerEvent(relayerToBeAdded);
+    }
+
+    function registerManagerAddRelayer(address relayer) external payable onlyNonRegisteredManager {
+        registerManager();
+        addRelayer(relayer);
+    }
+
+    function removeRelayer() external onlyRegisteredManager {
```
guagualvcha, line 175: I don't think there is a necessity to implement `removeRelayer`, we can do it by updating the relayer to a non-exist account.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,189 @@ import "./System.sol";
...
+        delete (managersAndRelayers[msg.sender]);
+
+        emit removeRelayerEvent(r);
+    }
+
+    function isRelayer(address relayerAddress) external override view returns (bool){
+        return relayerExistsMap[relayerAddress];
+    }
+
+    // TODO remove just for testing
```
guagualvcha, line 192: Overall, the name of the newly added function and variable, the field, is not that clear, especially since you introduce three-step to get a relayer, readability is not good.

contracts/RelayerHub.sol
emailtovamos: Could you tell the exact hardfork steps with regards to the changes in this PR? I was under the impression that when we change the BSC and deploy the new RelayerHub, the `init()` can be called again. 

Also how else can we ensure smoother transition so that the current relayers remain relayers right after the hardfork? By initialising the maps in a constructor?

contracts/RelayerHub.sol
emailtovamos: Ok cool. 
Should I get the rest relayers manually from [bscscan](https://bscscan.com/address/0x0000000000000000000000000000000000001006) (by checking which legitimate ones have registered but haven't unregistered) or is there any other better way?

contracts/RelayerHub.sol
guagualvcha: Overall  for the PR.

contracts/RelayerHub.sol
guagualvcha: It is better to have a way to not use a whitelist way to let them exit.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,206 @@ import "./System.sol";
...
+
+    uint256 public requiredDeposit;
+    uint256 public dues;
+
+    mapping(address => relayer) relayers; // old map holding the relayers which are to be allowed safe exit
+    mapping(address => bool) relayersExistMap;
+
+    mapping(address => manager) managers;
+    mapping(address => bool) managersRegistered;
+    mapping(address => bool) relayManagersExistMap;
```
bnb-twu, line 27: can't this be represented by one mapping? You could use `false` to denote manager hasn't registered, and `true` for a manger that has.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,206 @@ import "./System.sol";
...
+    mapping(address => address) managersAndRelayers;
```
bnb-twu, line 28:
```suggestion
    mapping(address => address) managerToRelayer;
```

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,206 @@ import "./System.sol";
...
+    mapping(address => bool) relayerExistsMap;
```
bnb-twu, line 29:
```suggestion
    mapping(address => bool) currentRelayers;
```

contracts/RelayerHub.sol
emailtovamos: Should I change it to `onlyApprovedManager` ? Simply `onlyManager` wouldn't be very specific since a manager can be registered or non-registered. 

Else feel free to suggest any other name that is less confusing.

contracts/RelayerHub.sol
emailtovamos: That way it won't be possible to know if a manager is added by Gov or not since solidity allows all keys to exist in a map with their `0`/`false` values (according to my understanding).

contracts/RelayerHub.sol
emailtovamos: Since `init()` or `construct()` can't be called again, I have [added](https://github.com/bnb-chain/bsc-genesis-contract/pull/205/commits/6dfa1690db3275431c080108c54e7571013fb988) the `update()` function which can only be called by a pre-decided address so that the code inside update() can be executed manually right after the deployment of the smart contract to ensure smooth transition. If this is acceptable then we can decide on the address. 

I also added the `exitOldRelayers()` for old relayers to call and exit.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,206 @@ import "./System.sol";
...
+
+    uint256 public constant INIT_REQUIRED_DEPOSIT = 1e20;
+    uint256 public constant INIT_DUES = 1e17;
+    address public constant WHITELIST_1 = 0xb005741528b86F5952469d80A8614591E3c5B632;
+    address public constant WHITELIST_2 = 0x446AA6E0DC65690403dF3F127750da1322941F3e;
+
+    uint256 public requiredDeposit;
+    uint256 public dues;
+
+    mapping(address => relayer) relayers; // old map holding the relayers which are to be allowed safe exit
```
guagualvcha, line 24: The white-list ones are safe, the rest are not sure.

guagualvcha, line 24: Will compile success? Since the structure of `relayer` is not defined.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,206 @@ import "./System.sol";
...
+    mapping(address => bool) relayersExistMap;
+
+    mapping(address => manager) managers;
+    mapping(address => bool) managersRegistered;
+    mapping(address => bool) relayManagersExistMap;
+    mapping(address => address) managerToRelayer;
+    mapping(address => bool) currentRelayers;
+
+    struct manager {
```
guagualvcha, line 33: The initial design needs to deposit 100BNB to become a relayer because we want to set a bar to relayers.

Now, a relayer can only be created through governance, I think this a higher bar than deposit 100BNB, so I think there is no need to deposit 100 BNB for newly added relayer now.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,206 @@ import "./System.sol";
...
-    emit paramChange(key, value);
-  }
-
-  function isRelayer(address sender) external override view returns (bool) {
-    return relayersExistMap[sender];
-  }
+contract RelayerHub is IRelayerHub, System, IParamSubscriber {
+    using SafeMath for uint256;
+
+    bool public alreadyUpdate;
```
guagualvcha, line 14: The whole storage structure will break if you add a new var here. For any upgrade smart contract, you can only append new variables, the order matters. (This is because of the smart contract storage layout)

guagualvcha: `alreadyUpdate` is a too generic name, since we upgrade this contract in the future, by then if we need add another `alreadyUpdate`, that will be confusing.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,206 @@ import "./System.sol";
...
+        emit relayerUnRegister(msg.sender);
+    }
+
+    function init() external onlyNotInit {
+        requiredDeposit = INIT_REQUIRED_DEPOSIT;
+        dues = INIT_DUES;
+        alreadyInit = true;
+    }
+
+    function update() external onlyAllowedParty {
```
guagualvcha, line 90: The `update` name is too generic, and I think `onlyAllowedParty` is not necessary, let us allow anyone to invoke it, there is no risk.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,199 @@ import "./System.sol";
...
+    bool public alreadyUpdate;
+
+    modifier onlyNonRegisteredManager() {
+        require(relayManagersExistMap[msg.sender], "manager does not exist");
+        _;
+    }
+
+    modifier onlyRegisteredManager() {
+        require(relayManagersExistMap[msg.sender], "manager does not exist");
+        require(managersRegistered[msg.sender], "manager not registered");
```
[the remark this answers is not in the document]
emailtovamos: You mean the first check is not needed since managers get registered later once they have been added by Gov?

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,199 @@ import "./System.sol";
...
+
+    mapping(address => relayer) relayers; // old map holding the relayers which are to be allowed safe exit
+    mapping(address => bool) relayersExistMap;
+
+    struct relayer {
+        uint256 deposit;
+        uint256 dues;
+    }
+
+    mapping(address => uint256) managerDeus;
```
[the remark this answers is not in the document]
emailtovamos: It won't work in the case where a manager is registered and all but his `Dues` are `0` unless I misunderstood what you suggested.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,199 @@ import "./System.sol";
...
+    }
+
+    function removeManagerByHimself() external {
+        // here the manager removes himself
+        removeManager(payable(msg.sender));
+    }
+
+    function removeManager(address payable managerAddress) internal {
+        // check if the manager address already exists
+        require(relayManagersExistMap[managerAddress], "manager doesn't exist");
```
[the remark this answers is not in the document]
emailtovamos: This check only exists once, the modifier checks for `msg.sender` whereas this checks for the address passed as parameter but called by the Gov.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,199 @@ import "./System.sol";
...
+    mapping(address => bool) relayersExistMap;
+
+    struct relayer {
+        uint256 deposit;
+        uint256 dues;
+    }
+
+    mapping(address => uint256) managerDeus;
+    mapping(address => bool) managersRegistered;
+    mapping(address => bool) relayManagersExistMap;
```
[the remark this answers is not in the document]
emailtovamos: A manager can exist without having its corresponding relayer. So just having `managerToRelayer` isn't enough for us to know if a manager exists.

contracts/RelayerHub.sol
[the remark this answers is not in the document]
emailtovamos: You mean a mapping where the key is `address` and value is the `struct`? There was a struct before which we got rid of 😄

contracts/RelayerHub.sol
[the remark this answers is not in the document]
emailtovamos: Almost the same meaning but not exactly. A manager can register himself and add a relayer and after some time can remove its relayer. So `managersRegistered` will still be `non-nil` but `managerToRelayer` will be `nil` for this manager address.

BTW My[ latest commit](https://github.com/bnb-chain/bsc-genesis-contract/pull/205/commits/286f4cebb905d182ab7fadc75356c3a8d16919b1) adds a missing check in this regard which should clarify things.

contracts/RelayerHub.sol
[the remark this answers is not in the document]
emailtovamos: Yes it used to have dues and deposit. If I make it a struct with bool and uint256, it will still remain a mapping right? The total number of maps won't be reduced. Or do you want to get rid of some other map?

contracts/RelayerHub.sol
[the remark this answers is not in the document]
emailtovamos: I added different flow for zero account and also added a test case which captured the bug.

contracts/RelayerHub.sol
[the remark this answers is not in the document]
emailtovamos: Done in [this commit](https://github.com/bnb-chain/bsc-genesis-contract/pull/205/commits/1e4ab6840810d5a04d02becae2b66c17959dbaad).

contracts/RelayerHub.template
```diff
@@ -8,87 +8,199 @@ import "./System.sol";
...
+    address public constant WHITELIST_1 = 0xA904540818AC9c47f2321F97F1069B9d8746c6DB;
+    address public constant WHITELIST_2 = 0x316b2Fa7C8a2ab7E21110a4B3f58771C01A71344;
+  {% elif network == 'QA' %}
+    address public constant WHITELIST_1 = 0x88cb4D8F77742c24d647BEf8049D3f3C56067cDD;
+    address public constant WHITELIST_2 = 0x42D596440775C90db8d9187b47650986E1063493;
+  {% elif network == 'testnet' %}
+    address public constant WHITELIST_1 = 0x9fB29AAc15b9A4B7F17c3385939b007540f4d791;
+    address public constant WHITELIST_2 = 0x37B8516a0F88E65D677229b402ec6C1e0E333004;
+  {% else %}
+    address public constant WHITELIST_1 = 0x04d63aBCd2b9b1baa327f2Dda0f873F197ccd186;
```
guagualvcha, line 26: This address seems wrong which does not match with mainnet setting.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,192 @@ import "./System.sol";
...
+
+        managerToRelayer[msg.sender] = relayerToBeAdded;
+        currentRelayers[relayerToBeAdded] = true;
+
+        emit updateRelayerEvent(oldRelayer, relayerToBeAdded);
+    }
+
+    function registerManagerAddRelayer(address r) external payable onlyNonRegisteredManager {
+        // register manager
+        managers[msg.sender] = manager(dues, true);
```
guagualvcha, line 185: A potential bug here:  You did not check the `msg.value == dues`, so that the msg value can be zero, but RelayerHub will transfer dues to system reward contract, where does the funds come?

guagualvcha: I will suggest no `dues` at all to make it simple.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,192 @@ import "./System.sol";
...
+
+    function whitelistInit() external {
+        require(!whitelistInitDone, "the whitelists already updated");
+        addInitRelayer(WHITELIST_1);
+        addInitRelayer(WHITELIST_2);
+        whitelistInitDone = true;
+    }
+
+    function addInitRelayer(address addr) internal {
+        managers[addr] = manager(dues, true);
```
guagualvcha, line 91: Why not do:
```
addManagerByGov(addr)
registerManagerAddRelayer(addr)
```
Much simple, and just need rename `addManagerByGov` to `addManager`

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,192 @@ import "./System.sol";
...
+        address oldRelayer = managerToRelayer[msg.sender];
+        delete currentRelayers[oldRelayer];
+
+        managerToRelayer[msg.sender] = relayerToBeAdded;
+        currentRelayers[relayerToBeAdded] = true;
+
+        emit updateRelayerEvent(oldRelayer, relayerToBeAdded);
+    }
+
+    function registerManagerAddRelayer(address r) external payable onlyNonRegisteredManager {
```
guagualvcha, line 183: If we dont have dues, then there is no need to have `registerManagerAddRelayer`,  just `updateRelayer` is enough.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,192 @@ import "./System.sol";
...
+    modifier exist() {
+        require(relayersExistMap[msg.sender], "relayer do not exist");
+        _;
+    }
+
+    event relayerRegister(address _relayer);
+    event relayerUnRegister(address _relayer);
+    event paramChange(string key, bytes value);
+
+    event removeManagerEvent(address _removedManager);
```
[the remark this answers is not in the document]
emailtovamos: Mainly because we can't have events and functions of the exact same name. But if an event is supposed to convey exactly what a function did then there's no reason to have them very separate core names. Therefore I added the "Event" suffix to distinguish. 

Otherwise for each of these 4 new events I will have to come up with similar sounding names as their corresponding functions. Previously there were `relayerRegister` event and `register` function for example. 

Is there any general best practice in this regard?

contracts/RelayerHub.sol
emailtovamos: It might require few additional changes if not an extra function since the `registerManagerAddRelayer()` function which will be removed and replaced by `updateRelayer()` and `updateRelayer()` can only be called by managers whereas with the current setup we are allowing anyone to call the `whitelistInit()` function to do the initialisation. 

So too much of a change to replace 3 lines of code. 😄

contracts/RelayerHub.sol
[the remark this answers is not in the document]
guagualvcha: There is a pattern to follow: 

For function, the name is like`doSomething`, for example, `registerRelayer`.

For event, the name is like `somethingDone`, like `relayerRegister` or `relayerUnRegistered`

guagualvcha (approved):

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,150 @@ import "./System.sol";
...
+    event addManagerByGovEvent(address _addedManager);
+    event registerManagerEvent(address _registeredManager);
+    event updateRelayerEvent(address _from, address _to);
+
+    function init() external onlyNotInit {
```
Pythonberg1997, line 55: We should not modify the init function. As the init function will never be called again.

Pythonberg1997 (approved):

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,149 @@ import "./System.sol";
...
+        emit managerAdded(managerToBeAdded);
+    }
+
+    // updateRelayer() can be used to add relayer for the first time, update it in future and remove it
+    // in case of removal we can simply update it to a non-existing account
+    function updateRelayer(address relayerToBeAdded) public onlyManager {
+        require(!isContract(relayerToBeAdded), "contract is not allowed to be a relayer");
+
+        address oldRelayer = managerToRelayer[msg.sender];
+        delete currentRelayers[oldRelayer];
```
cosinlink, line 136: This line should be moved to `line 145`. While `relayerToBeAdded == oldRelayer`, it will not fail to update as expected since `line 136`

cosinlink (changes requested):

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,149 @@ import "./System.sol";
...
+        }
+        emit paramChange(key, value);
+    }
+
+    function removeManagerByHimself() external {
+        // here the manager removes himself
+        removeManager(payable(msg.sender));
+    }
+
+    function removeManager(address payable managerAddress) internal {
```
cosinlink, line 104: why is the `managerAddress` payable ?

cosinlink (approved):

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,175 @@ import "./System.sol";
...
+        // emit success event
+        emit removeManagerAddress(managerAddress);
+        if (relayerAddress != address(0)) {
+            emit removeRelayer(relayerAddress);
+        }
+    }
+
+    function addManagerAddress(address managerToBeAdded) external onlyGov {
+        require(!relayManagersExistMap[managerToBeAdded], "manager already exists");
+
```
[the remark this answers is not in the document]
musk-j: Hi, @MatusKysel 
I am wondering if `!isContract` is helpful here.
My idea is, people might want to use smart contracts as `managers`, such as Multi-sig wallet.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,151 @@ import "./System.sol";
...
+    uint256 public constant INIT_REQUIRED_DEPOSIT = 1e20;
+    uint256 public constant INIT_DUES = 1e17;
+    address public constant WHITELIST_1 = 0xb005741528b86F5952469d80A8614591E3c5B632;
+    address public constant WHITELIST_2 = 0x446AA6E0DC65690403dF3F127750da1322941F3e;
+
+    uint256 public requiredDeposit; // have to keep it to not break the storage layout
+    uint256 public dues;
+
+    mapping(address => relayer) relayers; // old map holding the relayers which are to be allowed safe exit
```
musk-j, line 22: I think the variable name`relayers` is confusing here, if someone does not know the "smart contract upgrade" history.
Could you consider renaming it to `old_relayers` or `deprecated_relayers`.
If that is too compilcated, could we add a comment for `relayers`.

contracts/RelayerHub.sol
```diff
@@ -8,88 +8,151 @@ import "./System.sol";
...
-      require(false, "unknown param");
+contract RelayerHub is IRelayerHub, System, IParamSubscriber {
+    using SafeMath for uint256;
+
+    uint256 public constant INIT_REQUIRED_DEPOSIT = 1e20;
+    uint256 public constant INIT_DUES = 1e17;
+    address public constant WHITELIST_1 = 0xb005741528b86F5952469d80A8614591E3c5B632;
+    address public constant WHITELIST_2 = 0x446AA6E0DC65690403dF3F127750da1322941F3e;
+
+    uint256 public requiredDeposit; // have to keep it to not break the storage layout
```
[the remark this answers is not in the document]
emailtovamos: Ok, I didn't know making it `public` to `internal` won't affect storage layout. Thanks.

contracts/RelayerHub.sol
emailtovamos: hmm I agree. Just to confirm, changing name doesn't affect storage layout right?

contracts/RelayerHub.sol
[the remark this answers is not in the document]
emailtovamos: How likely is this to happen though? Especially considering Account Abstraction isn't active. 
Also do we know why exactly contracts weren't allowed before?

test/RelayerHub.t.sol
```diff
@@ -3,50 +3,430 @@ pragma solidity ^0.8.10;
...
+        assertTrue(newRelayerHub.isProvisionalRelayer(newRelayer));
+
+        address randomManager = payable(addrSet[addrIdx++]);
+        vm.prank(newRelayer, newRelayer);
+        vm.expectRevert("provisional is not set for this manager");
+        newRelayerHub.acceptBeingRelayer(randomManager);
+
+    }
+
+    //  function testCannotRegister() public {
```
Mister-EA, line 406: Is there a reason this is commented out?

generate-relayerhub.js
```diff
@@ -23,6 +23,7 @@ program.parse(process.argv);
 
 const data = {
   mock: program.mock,
+  network: program.network,
```
emailtovamos, line 26: We also should delete `network` from `generate-system.js` as it is no longer used there.

PR #212, opened 2022-12-07
[r4r] fix(BEP-171): fix bug that SLASH_CHANNELID cannot be suspended in order for consensus not to be affected

cosinlink:
### Description

This is a fix for BEP-171, make SLASH_CHANNELID cannnot be suspended in order for consensus not to be affected 

### Rationale

This is a very serious bug that can affect the parlia consensus.
In the previous design,  all channels will be closed and cannot cross-chain communicate once it is suspended. This will cause slash to fail and the BSC nodes will not run properly.

### Changes

Notable changes: 
* SLASH_CHANNELID will be suspended


yutianwu (approved):

contracts/SlashIndicator.sol
```diff
@@ -96,9 +98,13 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
     }
     indicator.height = block.number;
     if (indicator.count % felonyThreshold == 0) {
-      indicator.count = 0;
-      IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).felony(validator);
-      ICrossChain(CROSS_CHAIN_CONTRACT_ADDR).sendSynPackage(SLASH_CHANNELID, encodeSlashPackage(validator), 0);
+      try ICrossChain(CROSS_CHAIN_CONTRACT_ADDR).sendSynPackage(SLASH_CHANNELID, encodeSlashPackage(validator), 0) {
+          indicator.count = 0;
+          IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).felony(validator);
```
guagualvcha, line 103: Even the channel is closed, we still do `          IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).felony(validator);`.

contracts/SlashIndicator.sol
```diff
@@ -96,9 +98,13 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
...
+      } catch (bytes memory reason) {
+          indicator.count--;
```
guagualvcha, line 105: Because it is punished, so no need to count down.

guagualvcha (approved):

Pythonberg1997 (approved):

contracts/SlashIndicator.sol
cosinlink: fixed

PR #221, opened 2022-12-19
docs: add how to generate mainnet/testnet/QA genesis file

guagualvcha:
### Description

In this pr, the readme guide user to generate the genesis file in one short command.

### Rationale
Currently, when we upgrade a smart contract on BSC network, it is hard for another developer to verify whether the byte code is right or not.

We introduce some docs in this PR, so that reviewer can easily verify the bytecode.

### Example

NA

### Changes
Nono

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,17 @@ Truffle: v5.1.31
 
 Solc: 0.6.4+commit.1dca32f3
 
+Tips: You can manage multi version of Solc and Node:
+```Shell
+## Install solc-select and solc
+pip3 install solc-select
+solc-select install 0.6.4 && solc-select use 0.6.4
+
+## Install nvm and node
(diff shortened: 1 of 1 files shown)


forcodedancing (approved):

README.md
```diff
@@ -24,6 +24,10 @@ Truffle: v5.1.31
 
 Solc: 0.6.4+commit.1dca32f3
 
+Tips: You can manage multi version of 
+- solc by [solc-select](https://github.com/crytic/solc-select).
```
emailtovamos, line 28: Could we add an example command e.g. `solc-select install 0.6.4 && solc-select use 0.6.4` ? 
Similar for nvm.

guagualvcha: sure, updated.

PR #262, opened 2023-04-17
fix: slash more bnbs for malicious vote and no matter whether slashed validator is active.

NathanBSC:
Description

malicious vote affects the safety of bsc seriously, 
so we need slash more BNBs for malicious vote and no matter whether slashed validator is active.

Rationale
1. after a validator generating a malicious vote, it can go down on purpose to avoid to be slashed.
	so we change the mechanism to slash it no matter whether it's active.
2. slash amount is 50 BNBs, more need to be slashed.
	so we send a different crosschain pack and handle it in bc node correspondingly 

Example
tested in new qa
N/A

### DIFF
--- a/contracts/SlashIndicator.sol
+++ b/contracts/SlashIndicator.sol
@@ -37,6 +37,7 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
   uint256 public constant INIT_FINALITY_SLASH_REWARD_RATIO = 20;
 
   uint256 public finalitySlashRewardRatio;
+  uint256 public enableMaliciousVoteSlash;
 
   event validatorSlashed(address indexed validator);
   event maliciousVoteSlashed(bytes32 indexed voteAddrSlice);
@@ -197,6 +198,7 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
   }
 
   function submitFinalityViolationEvidence(FinalityEvidence memory _evidence) public onlyInit onlyRelayer {
+    require(enableMaliciousVoteSlash > 0, "malicious vote slash not enabled");
     if (finalitySlashRewardRatio == 0) {
       finalitySlashRewardRatio = INIT_FINALITY_SLASH_REWARD_RATIO;
     }
@@ -304,6 +306,9 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
       uint256 newFinalitySlashRewardRatio = BytesToTypes.bytesToUint256(32, value);
       require(newFinalitySlashRewardRatio >= 10 && newFinalitySlashRewardRatio < 100, "the finality slash reward ratio out of range");
       finalitySlashRewardRatio = newFinalitySlashRewardRatio;
+    } else if (Memory.compareStrings(key, "enableMaliciousVoteSlash")) {
+      require(value.length == 32, "length of enableMaliciousVoteSlash mismatch");
+      enableMaliciousVoteSlash = BytesToTypes.bytesToUint256(32, value);
     } else {
       require(false, "unknown param");
     }
--- a/contracts/SlashIndicator.sol
+++ b/contracts/SlashIndicator.sol
@@ -39,6 +39,7 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
   uint256 public finalitySlashRewardRatio;
 
   event validatorSlashed(address indexed validator);
+  event maliciousVoteSlashed(bytes32 indexed voteAddrSlice);
   event indicatorCleaned();
   event paramChange(string key, bytes value);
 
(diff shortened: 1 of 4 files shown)


contracts/SlashIndicator.sol
```diff
@@ -213,27 +214,26 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
...
-    require(verifyBLSSignature(_evidence.voteA, _evidence.voteAddr) &&
-      verifyBLSSignature(_evidence.voteB, _evidence.voteAddr), "verify signature failed");
-
-    uint256 amount = (address(SYSTEM_REWARD_ADDR).balance * finalitySlashRewardRatio) / 100;
-    ISystemReward(SYSTEM_REWARD_ADDR).claimRewards(msg.sender, amount);
-    IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).felony(valAddr);
-    ICrossChain(CROSS_CHAIN_CONTRACT_ADDR).sendSynPackage(SLASH_CHANNELID, encodeSlashPackage(valAddr), 0);
-    emit validatorSlashed(valAddr);
+    // send slash msg to bc
+    ICrossChain(CROSS_CHAIN_CONTRACT_ADDR).sendSynPackage(SLASH_CHANNELID, encodeVoteSlashPackage(_evidence.voteAddr), 0);
```
NathanBSC, line 232: try...catch is need here?

NathanBSC:
double sign 100BNB, not 1000?

contracts/SlashIndicator.sol
```diff
@@ -47,6 +48,7 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
   event crashResponse();
 
   event failedFelony(address indexed validator, uint256 slashCount, bytes failReason);
+  event failedVoteSlash(bytes32 indexed voteAddrSlice, bytes failReason);
```
guagualvcha, line 51: better rename to `maliciousVoteSlash`

contracts/SlashIndicator.sol
```diff
@@ -213,27 +215,30 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
...
     (address[] memory vals, bytes[] memory voteAddrs) = IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).getLivingValidators();
-    address valAddr;
-    bytes memory voteAddr = _evidence.voteAddr;
     for (uint i; i < voteAddrs.length; ++i) {
-      if (BytesLib.equal(voteAddrs[i],  voteAddr)) {
-        valAddr = vals[i];
+      if (BytesLib.equal(voteAddrs[i],  _evidence.voteAddr)) {
+        uint256 amount = (address(SYSTEM_REWARD_ADDR).balance * finalitySlashRewardRatio) / 100;
+        ISystemReward(SYSTEM_REWARD_ADDR).claimRewards(msg.sender, amount);
+        IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).felony( vals[i]);
```
guagualvcha, line 227: In the common case, both `felony` and `VoteSlash` will trigger, so two cross chain package will be trigger, I think this need to optimized.

guagualvcha:
Please add some UTs and also refine the PR title and description.

contracts/SlashIndicator.sol
```diff
@@ -213,27 +215,30 @@ contract SlashIndicator is ISlashIndicator,System,IParamSubscriber, IApplication
...
-      verifyBLSSignature(_evidence.voteB, _evidence.voteAddr), "verify signature failed");
 
-    uint256 amount = (address(SYSTEM_REWARD_ADDR).balance * finalitySlashRewardRatio) / 100;
-    ISystemReward(SYSTEM_REWARD_ADDR).claimRewards(msg.sender, amount);
-    IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).felony(valAddr);
-    ICrossChain(CROSS_CHAIN_CONTRACT_ADDR).sendSynPackage(SLASH_CHANNELID, encodeSlashPackage(valAddr), 0);
-    emit validatorSlashed(valAddr);
+    // send slash msg to bc
+    bytes32 voteAddrSlice = BytesLib.toBytes32(_evidence.voteAddr,0);
+    try ICrossChain(CROSS_CHAIN_CONTRACT_ADDR).sendSynPackage(SLASH_CHANNELID, encodeVoteSlashPackage(_evidence.voteAddr), 0) {
```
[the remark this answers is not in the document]
NathanBSC: 1. when  a slash generates, but fails,
   we should have a way to know this, emit a event is a easy way
2. for SLASH_CHANNELID, try... catch also added in following place
image
3. sendSynPackage-->sendPackage
    sendPackage has modifier 'whenNotSuspended'
    may not be satisfied after suspend() is called
   image
image
image

contracts/SlashIndicator.sol
NathanBSC: accepted
failedMaliciousVoteSlash

guagualvcha (approved):

Pythonberg1997 (approved):

PR #290, opened 2023-05-07
fix: protect validator from being slashed for malicious vote signed b…

NathanBSC:
## Description

for now,
validator can be slashed for malicious vote signed by verify old vote keys.
this PR limit that vote keys should be in last two validator set which come from breathe block.

## Changes

Notable changes:
* ...

### DIFF
--- a/contracts/BSCValidatorSet.sol
+++ b/contracts/BSCValidatorSet.sol
@@ -83,6 +83,8 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
   uint256 public finalityRewardRatio;
   uint256 public previousHeight;
   uint256 public previousBalanceOfSystemReward;
+  bytes[] public previousVoteAddrFullSet;
+  bytes[] public currentVoteAddrFullSet;
 
   struct Validator {
     address consensusAddress;
@@ -803,6 +805,10 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
       }
     }
 
+    // update vote addr full set
+    setPreviousVoteAddrFullSet();
+    setCurrentVoteAddrFullSet();
+
     // make sure all new validators are cleared maintainInfo
     // should not happen, still protect
     numOfMaintaining = 0;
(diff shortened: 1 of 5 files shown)


contracts/BSCValidatorSet.sol
```diff
@@ -83,6 +83,7 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
   uint256 public finalityRewardRatio;
   uint256 public previousHeight;
   uint256 public previousBalanceOfSystemReward;
+  bytes[] public previousVoteAddrsSet;
```
guagualvcha, line 86: `previousVoteAddrsSet` typo? Get warning in my IDE

NathanBSC: > `previousVoteAddrsSet` typo? Get warning in my IDE

previousVoteAddrsSet -->previousVoteAddrSet ?
but all with warning , because it's not a existed word?

NathanBSC:
two commits, need squash now

contracts/BSCValidatorSet.sol
```diff
@@ -838,6 +841,48 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     return voteAddrs;
   }
 
+  function setpreviousVoteAddrSet() private {
+    uint n = previousVoteAddrSet.length;
+    uint m = validatorExtraSet.length;
```
guagualvcha, line 846: seems wrong? as the `validatorExtraSet ` is still old.

contracts/BSCValidatorSet.sol
```diff
@@ -838,6 +841,48 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
+  }
+
+  function isMonitoredForMalicousVote(bytes calldata voteAddr) external override view returns (bool) {
+    uint m = validatorExtraSet.length;
+    for (uint i; i<m; ++i) {
+      if (BytesLib.equal(voteAddr, validatorExtraSet[i].voteAddress)) {
+        return true;
+      }
+    }
+
```
guagualvcha, line 875: `validatorExtraSet` and `previousVoteAddrSet` may not match, as the order is not the exact same.

contracts/BSCValidatorSet.sol
NathanBSC: it‘s by design

guagualvcha (approved):

Pythonberg1997 (approved):

PR #371, opened 2023-08-21
Support local deployment

Mister-EA:
This PR facilitates the generation of the genesis contracts for a local deployment of a BSC network. 

The  [node-deploy](https://github.com/bnb-chain/node-deploy/) repository currently modifies the genesis contract templates using `sed` which is prone to error and not easy to maintain due to overwriting the original files, but with this PR the genesis configurations for a local network can be generated simply by including the flag `--network local` .

### Motivation
To run a local network several additional steps need to be coordinated to enable recently added features: e.g. initializing the whitelisted relayers, enabling the finality reward,  enabling malicious vote slashing etc.  These steps are normally performed through standard transactions or through governance, but the flows may be complicated especially in a local environment. 

To facilitate these steps, the initialization is done in the contract code itself when generating the bytecode for the genesis contracts. This PR ensures that these initializations happen automatically when passing the `--network local` flag. (i.e. `node generate-genesis.js --network local`. 

### Additional improvements
- Automatically calculating the hexadecimal `bscChainId` value  from the  `chainid` flag.
- Ability to define custom whitelist relayer addresses for local deployment.
- Ability to initialize the initial BNB holders by running `node generate-initHolders.js --initHolders <commaSeparatedAddresses>`


contracts/RelayerHub.template
```diff
@@ -73,6 +73,9 @@ contract RelayerHub is IRelayerHub, System, IParamSubscriber {
     function init() external onlyNotInit {
         requiredDeposit = INIT_REQUIRED_DEPOSIT;
         dues = INIT_DUES;
+        {% if network == 'mock' %}
```
buddh0, line 76: mock？ or local？

buddh0:
maybe we need delete bscChaindId in file package.json

contracts/RelayerHub.template
Mister-EA: Good catch! It should be local

buddh0 (approved):

init_holders.js
```diff
@@ -1,15 +1,19 @@
...
+  },
+  {
+     address: "0x6c468CF8c9879006E22EC4029696E005C2319C9D",
+     balance: web3.utils.toBN("500000000000000000000").toString("hex") // 500e18
+  },
+  
+  {
+     address: "0x04d63aBCd2b9b1baa327f2Dda0f873F197ccd186",
+     balance: web3.utils.toBN("500000000000000000000000000").toString("hex") // 500000000e18
+  },
```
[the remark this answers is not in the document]
Mister-EA: Not sure, they were already there in the template. You can add your own addresses to the list via the template, and manage the keys yourself.

PR #430, opened 2023-12-05
feat: add apy and reward record of validators

Pythonberg1997:


contracts/BC_fusion/StakeHub.sol
```diff
@@ -921,4 +955,18 @@ contract StakeHub is System, Initializable {
             emit ValidatorJailed(valInfo.operatorAddress);
         }
     }
+
+    function _truncUint256(uint256 x) internal pure returns (uint256) {
```
zjubfd, line 959: I believe the code here is not invented by you, what is the reference?

contracts/BC_fusion/StakeHub.sol
```diff
@@ -921,4 +955,18 @@ contract StakeHub is System, Initializable {
...
+        uint256 leadingZero;
+        uint256 mask = 1 << 255;
+        for (uint256 i; i < 136; ++i) {
```
zjubfd, line 962: for loop will cost too much gas here?

contracts/BC_fusion/StakeHub.sol
```diff
@@ -712,6 +721,31 @@ contract StakeHub is System, Initializable {
...
+    /**
+     * @notice the returned reward may be not accurate, it's been truncated
+     * @return the validator's reward and apy of the day(with 2 decimals)
+     */
+    function getValidatorRewardInfo(
+        address operatorAddress,
+        uint256 dayIndex
+    ) external view returns (uint256, uint256) {
+        uint256 info = _rewardInfo[operatorAddress][dayIndex];
+        uint256 mask = (1 << 120) - 1;
```
zjubfd, line 733: I dont mind to store one more uint256 instead of so many tricky calculation here, because break block is very rare.

zjubfd (approved):

PR #433, opened 2023-12-06
feat: add python script

Pythonberg1997:


scripts/init_holders.template
```diff
@@ -1,17 +1,9 @@
-const web3 = require('web3')
+const web3 = require('web3');
 const init_holders = [
-  {
```
zjubfd, line 3: Can we still keep these two address, becausee the default relayer account on dev-net and QA are these two.

Pythonberg1997: Sure. I was just wondering where those two addresses came from.

PR #440, opened 2023-12-11
chore: update storages and generate.py

Pythonberg1997:
### Description

This pr is mainly to update generate script for modifying the init setting of some contracts for dev net.

### Rationale

For test convenience

### Changes

Notable changes: 
* Update generate script
* Update some contracts' storages


contracts/BC_fusion/interface/IStakeHub.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity 0.8.17;
 interface IStakeHub {
     function DEAD_ADDRESS() external view returns (address);
     function INIT_LOCK_AMOUNT() external view returns (uint256);
```
zjubfd, line 6: If `Lock Amount` is a const, I think no need to start with `INIT`.

zjubfd (approved):

PR #496, opened 2024-02-15
chore: fix audit issues

Pythonberg1997:
### Description

This pr is to fix some issues from audit reports

### DIFF
--- a/contracts/BSCValidatorSet.sol
+++ b/contracts/BSCValidatorSet.sol
@@ -930,6 +930,7 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
         validatorExtraSet[i].isMaintaining = false;
         validatorExtraSet[i].enterMaintenanceHeight = 0;
       } else {
+        currentValidatorSet[i].votingPower = newValidatorSet[i].votingPower;
         // update the vote address if it is different
         if (!BytesLib.equal(newVoteAddrs[i], validatorExtraSet[i].voteAddress)) {
           validatorExtraSet[i].voteAddress = newVoteAddrs[i];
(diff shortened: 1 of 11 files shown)


contracts/BSCValidatorSet.sol
```diff
@@ -1122,6 +1121,12 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     address validator;
     bool isFelony;
 
+    for (uint i; i<_validatorSet.length; ++i) {
```
zjubfd, line 1125:
```
      for (uint k; k<_validatorSet.length; ++k) {
        if (_validatorSet[k].consensusAddress == validator) {
          _validatorSet[k].jailed = true;
          ++numOfFelony;
          break;
        }
      }
```
because of above logic, it is possible that one `_validatorSet[i]` cause `numOfFelony` increase twice.

contracts/BSCValidatorSet.sol
```diff
@@ -967,7 +968,7 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
    * Vote address is not considered
    */
   function isSameValidator(Validator memory v1, Validator memory v2) private pure returns(bool) {
-    return v1.consensusAddress == v2.consensusAddress && v1.feeAddress == v2.feeAddress && v1.BBCFeeAddress == v2.BBCFeeAddress && v1.votingPower == v2.votingPower;
+    return v1.consensusAddress == v2.consensusAddress && v1.feeAddress == v2.feeAddress && v1.BBCFeeAddress == v2.BBCFeeAddress;
```
zjubfd, line 971: Can you ensure that if we remove this check, the voting power of validator can update correctly?

contracts/BSCValidatorSet.sol
Pythonberg1997: Fixed

contracts/BSCValidatorSet.sol
Pythonberg1997: Yes. There's only 1 call of `isSameValidator`.

zjubfd (approved):

PR #499, opened 2024-02-18
fix: improve the code readablity

zjubfd:
The current implementation is not easy to understand because `numOfFelony` will increase in two different place.

After change, we count the number of felony validators just before we use it to generate the slice.

### DIFF
--- a/contracts/BSCValidatorSet.sol
+++ b/contracts/BSCValidatorSet.sol
@@ -1122,13 +1122,6 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
     address validator;
     bool isFelony;
 
-    // count the number of felony validators before forcing maintaining validators exit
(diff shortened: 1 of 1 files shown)


contracts/BSCValidatorSet.sol
```diff
@@ -1154,15 +1147,19 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
...
-            _validatorSet[k].jailed = true;
-            ++numOfFelony;
-          }
+          _validatorSet[k].jailed = true;
           break;
         }
       }
     }
 
+    // count the number of felony validators before forcing maintaining validators exit
```
Pythonberg1997, line 1156: This comment is incorrect

Pythonberg1997 (approved):

contracts/BSCValidatorSet.sol
zjubfd: fixed

PR #565, opened 2024-06-24
fix: array out-of-bounds issue in `_exitMaintenance`

Pythonberg1997:
### Description

This pr is to fix an issue in `_exitMaintenance` that the array index could be out-of-bounds 

### Rationale

### Example

### Changes

Notable changes:
* ...


contracts/BSCValidatorSet.sol
```diff
@@ -1294,9 +1298,10 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
             _misdemeanor(validator);
         }
 
-        // step 3: modify global storage
-        --numOfMaintaining;
-        validatorExtraSet[index].isMaintaining = false;
+        // step 4: clear isMaintaining info
```
unclezoro, line 1301: if this is the case, we can move `validatorExtraSet[index].isMaintaining = false;` up without condition.

contracts/BSCValidatorSet.sol
```diff
@@ -1202,7 +1202,7 @@ contract BSCValidatorSet is IBSCValidatorSet, System, IParamSubscriber, IApplica
             validator = currentValidatorSet[i].consensusAddress;
 
             // exit maintenance
-            isFelony = _exitMaintenance(validator, i, miningValidatorCount);
+            isFelony = _exitMaintenance(validator, i, miningValidatorCount, false);
```
unclezoro, line 1205: should't it be true?

contracts/interface/ISlashIndicator.sol
```diff
@@ -2,7 +2,7 @@ pragma solidity 0.6.4;
 
 interface ISlashIndicator {
     function clean() external;
-    function downtimeSlash(address validator, uint256 count) external;
```
unclezoro, line 5: I will suggest do not include this in the event `shouldRevert`

contracts/BSCValidatorSet.sol
Pythonberg1997: System tx should be false.

contracts/interface/ISlashIndicator.sol
Pythonberg1997: This is not an event. It's the function interface

zzzckck (approved):

zzzckck:
pls open an issue to track the failed UT, better to bring it up.

NathanBSC (dismissed):

cosinlink (approved):

unclezoro (approved):

PR #647, opened 2025-09-19
feat: change value scope for penalty params

allformless:
### Description

feat: change value scope for penalty params

### Rationale

### Example

### Changes

Notable changes:
* ...


buddh0:
I'm sure the failure of UT has no relation with this PR. 
so ignore it for now.

zzzckck (approved):

PR #664, opened 2026-05-07
fix: propagate slash eviction to BSCValidatorSet after consensus key rotation

zlacfzy:
### Description

Fix slashing pipeline cross-contract inconsistency. After a consensus-key rotation, `StakeHub._jailValidator` did not propagate eviction to `BSCValidatorSet`, and `SlashIndicator.felony(K_old)` could silently miss because `BSCValidatorSet.currentValidatorSetMap` had been rebuilt to `K_new` at the previous breathe-block. The slashed validator could keep mining for up to ~24h until the next election dropped it via `votingPower=0`.

This PR adds `IBSCValidatorSet.felony(valInfo.consensusAddress)` after `_jailValidator` in `doubleSignSlash` and `maliciousVoteSlash`, evicting via `K_new`. SlashIndicator's existing `felony(K_old)` calls are retained to cover the pre-breathe-block window. `BSCValidatorSet.felony` is documented as idempotent so belt-and-suspenders is safe

### Rationale

### Example

### Changes

- `StakeHub.sol::doubleSignSlash`: cross-call `BSCValidatorSet.felony(valInfo.consensusAddress)`.
- `StakeHub.sol::maliciousVoteSlash`: same cross-call.
- `BSCValidatorSet.sol::felony`: add NatSpec on idempotency.

### DIFF
--- a/contracts/StakeHub.sol
+++ b/contracts/StakeHub.sol
@@ -710,6 +710,9 @@ contract StakeHub is SystemV2, Initializable, Protectable {
         if (!canSlash) revert AlreadySlashed();
         uint256 slashAmount = IStakeCredit(valInfo.creditContract).slash(felonySlashAmount);
         _jailValidator(valInfo, jailUntil);
+        // Evict using the post-rotation consensus key; SlashIndicator's voteAddr-based
+        // eviction covers the case where BSCValidatorSet has not yet synced K_new.
+        IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).felony(valInfo.consensusAddress);
 
         emit ValidatorSlashed(operatorAddress, jailUntil, slashAmount, SlashType.MaliciousVote);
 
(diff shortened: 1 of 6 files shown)


allformless (dismissed):

allformless (dismissed):

contracts/StakeHub.sol
unclezoro, line 730: this function can only call by slash contract, while slash contract already do the felony call here: 

https://github.com/bnb-chain/bsc-genesis-contract/blob/34618f607f8356cf147dde6a69fae150bd53d5bf/contracts/SlashIndicator.sol#L262

https://github.com/bnb-chain/bsc-genesis-contract/blob/34618f607f8356cf147dde6a69fae150bd53d5bf/contracts/SlashIndicator.sol#L304

zlacfzy: The slash contract calls felony with an evidence-derived key — the signer (signing-time key) for double-sign, the voteAddr-matched living-set key for malicious-vote.
After a consensus-key rotation that key is stale: currentValidatorSetMap[K_old] is 0, so felony no-ops and the slashed+jailed validator stays active.
The new call uses valInfo.consensusAddress (current K_new), covering exactly that window. In the non-rotation case it's a harmless idempotent no-op, same pattern as _checkValidatorSelfDelegation (StakeHub.sol:1239).

unclezoro (approved):

pythonberg1997 (approved):
LGTM. And maybe add a rotation regression test?

buddh0 (approved):

allformless (approved):

zzzckck (approved):

PR #686, opened 2026-08-31
fix: evict the active-set key on slash after rotation, and pay the finality reporter unconditionally

zlacfzy:
### Description

Two coupled fixes on the slashing/eviction path, kept in one PR because the second is only correct once the first is in place.

1. **Evict the active-set consensus key on slash after a rotation.** `BSCValidatorSet.felony(addr)` no-ops when `addr` is not in the active set. Between `editConsensusAddress` (which updates only StakeHub's `consensusAddress` to K_new) and the next breathe-block resync, the active set is still keyed on K_old, so a slash that evicts only the current key can miss. StakeHub now records the pre-rotation key per operator and, in every StakeHub-driven felony (`doubleSignSlash`, `maliciousVoteSlash`, and the self-delegation-floor path), evicts both the current and the pre-rotation key via `_felonyActiveKey`. `felony()` on the non-active key is a harmless no-op, so whichever key is actually seated is removed regardless of how the slash resolved.

2. **Pay the finality-violation reporter unconditionally.** `submitFinalityViolationEvidence` previously paid the reporter only inside a loop that matched the evidence vote address against a CURRENT living validator's key, so a validator that rotated its vote key between the offence and the report was fully slashed while the reporter got nothing. It now calls `maliciousVoteSlash` first (which resolves the operator from the vote address, evicts it, and reverts the whole submission on already-slashed/expired) and then pays the reporter unconditionally, mirroring `submitDoubleSignEvidence`.

### Rationale

Fix 2 removes the loop's `felony(vals[i])`, which in the pre-resync consensus-rotation window was the only call that targeted the active-set key K_old; the remaining `maliciousVoteSlash` eviction targets K_new and would no-op there. Fix 1 makes `maliciousVoteSlash` evict both keys, so removing that loop is safe. Shipping them together avoids a window where fix 2 alone would weaken malicious-vote eviction.

### Changes

- `StakeHub`: record the pre-rotation consensus address in `editConsensusAddress`; add `_felonyActiveKey` and use it from `doubleSignSlash`, `maliciousVoteSlash`, and `_checkValidatorSelfDelegation` (new mapping appended to storage, upgrade-safe).
- `SlashIndicator`: in `submitFinalityViolationEvidence`, slash first then pay the reporter unconditionally; remove the vestigial current-living-key loop.
- `test`: slash-after-rotation eviction regression.

### DIFF
--- a/contracts/SlashIndicator.sol
+++ b/contracts/SlashIndicator.sol
@@ -252,15 +252,10 @@ contract SlashIndicator is ISlashIndicator, System, IParamSubscriber, IApplicati
             "verify signature failed"
         );
 
-        // Slash + jail + evict the offender (StakeHub resolves the operator from the vote
-        // address and evicts it). If it reverts (e.g. already slashed / expired), the whole
-        // submission reverts and no reward is paid.
+        // Slash/jail/evict first (reverts on already-slashed/expired, blocking any reward),
+        // then reward the reporter unconditionally as submitDoubleSignEvidence does. The old
+        // living-key loop skipped the reward whenever the vote key had been rotated.
         IStakeHub(STAKE_HUB_ADDR).maliciousVoteSlash(_evidence.voteAddr);
-
-        // Reward the reporter unconditionally, mirroring submitDoubleSignEvidence. The prior
-        // implementation only paid when the vote address matched a CURRENT living validator's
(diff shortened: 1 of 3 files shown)


allformless (approved):

will-2012 (commented):
LGTM.

joey0612 (approved):

PR #687, opened 2026-08-31
fix: cap validator description field length

zlacfzy:
### Description

Bound the length of a validator's free-form description fields (`identity`, `website`, `details`) in `createValidator` and `editDescription`.

### Rationale

The daily breathe-block validator-set update reads the whole registry through `StakeHub.getValidatorElectionInfo`, which copies each validator's full struct — including these strings — into memory. The fields have no length bound and the registry is append-only, so the per-entry cost of that enumeration is attacker-controlled. Capping each field keeps the election read's cost bounded so the registry cannot be grown in a way that makes `updateValidatorSetV2` fail on the node's `eth_call` gas cap. `moniker` is already length-checked by `_checkMoniker`.

### Example

`createValidator`/`editDescription` now revert with `InvalidDescription` when `identity`, `website`, or `details` exceeds `MAX_DESCRIPTION_LENGTH` (280 bytes). Existing stored descriptions are unaffected; only new create/edit calls are validated.

### Changes

- `StakeHub`: add `MAX_DESCRIPTION_LENGTH` and `_checkDescriptionLength`, called from `createValidator` and `editDescription`; add `InvalidDescription` error.
- `test`: regression covering at-cap acceptance and over-cap rejection on both paths.

### DIFF
--- a/contracts/StakeHub.sol
+++ b/contracts/StakeHub.sol
@@ -29,6 +29,12 @@ contract StakeHub is SystemV2, Initializable, Protectable {
 
     uint256 public constant INIT_MAX_NUMBER_NODE_ID = 5;
 
+    // Max byte length of each free-form description field (identity/website/details).
+    // Bounds the per-validator metadata that the daily election read copies out of storage,
+    // so an append-only registry of large descriptions cannot push getValidatorElectionInfo
+    // over the node's eth_call gas cap and halt the breathe-block validator-set update.
+    uint256 public constant MAX_DESCRIPTION_LENGTH = 280;
+
     // receive fund status
     uint8 private constant _DISABLE = 0;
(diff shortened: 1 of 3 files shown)


contracts/StakeHub.sol
```diff
@@ -52,6 +56,7 @@ contract StakeHub is SystemV2, Initializable, Protectable {
     error InvalidCommission();
     // @notice signature: 0x5dba5ad7
     error InvalidMoniker();
+    error InvalidDescription();
```
will-2012, line 59: Add notice signature 0x...?

will-2012 (commented):
check the failed actions? need reproduce genesis

allformless (approved):

joey0612 (approved):

PR #694, opened 2026-09-14
feat: implement BEP-714 early validator maintenance

zlacfzy:
### Description

Implement [BEP-714](https://github.com/bnb-chain/BEPs/pull/714): when `slash()` raises a validator's missed-block count to exactly `maintenanceThreshold` (default 40, governable), the validator contract admits it to maintenance without penalty through the same path a misdemeanor already uses. Maintenance settlement is unchanged from BEP-127. This is the system-contract portion of the Jenner upgrade; activation remains unscheduled.

### Rationale

Remove unavailable validators from rotation before any economic penalty, with the smallest possible change: no new system transaction, no new accounting. A validator refused at the threshold miss (capacity full, daily allowance used, maintenance disabled) is not retried; the misdemeanor path remains the next automatic entry.

### Example

A validator is admitted on its 40th miss with its income intact; misses after entry are not counted. Exiting after 199 equivalent maintenance misses settles nothing and ordinary slashing resumes at 41.

```bash
forge test --match-contract BEP714Test
```

### Changes

- `contracts/SlashIndicator.sol`: governable `maintenanceThreshold` (default 40, `0 < maintenanceThreshold < misdemeanorThreshold`); `slash()` gains one branch that calls `tryEnterMaintenance` on the threshold miss.
- `contracts/BSCValidatorSet.sol`: `tryEnterMaintenance` (onlySlash, no penalty, no revert when ineligible); `canEnterMaintenance` stops at the first other working validator instead of walking the set twice through `getValidators()`, so an admitting slash stays well inside the client's 1M system-transaction reserve.
- Interface, ABIs, and Validator/Slash genesis bytecode for mainnet, chapel and rialto.
- `test/BEP714.t.sol`: six tests: threshold admission without penalty, capacity checked only at the threshold miss with the misdemeanor path as fallback, disabled maintenance, last working validator, BEP-127 settlement untouched, governance bounds and authorization.

`maintainSlashScale=3`, `misdemeanorThreshold=200`, `felonyThreshold=600`, and `turnLength=4` remain separate governance changes. Both runtimes must be installed in the same fork block: with only SlashIndicator upgraded, the threshold slash reverts.

### Validation

- Local Foundry: 6 tests passed.
- Genesis regenerated per network from clean sources; only `0x1000` and `0x1001` differ from `develop`. The unrelated StakeHub drift produced by the testnet and dev generators is left untouched.
- The companion client's embedded-bytecode test (`TestJennerMaintenanceContractPair`) admits on the 40th slash in a 45-validator set and bounds the slash's cold gas.
- The GitHub unit-test job fails before tests start because Anvil at `127.0.0.1:8545` refuses connections; unrelated to this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


will-2012 (commented):
LGTM.

will-2012 (commented):
LGTM.

contracts/SlashIndicator.sol
```diff
@@ -121,17 +125,40 @@ contract SlashIndicator is ISlashIndicator, System, IParamSubscriber, IApplicati
...
+        }
+        // the validator contract snapshots the count when it enters maintenance below, so store it first
+        indicators[validator] = indicator;
+        if (isFelony) {
             IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).felony(validator);
             _downtimeSlash(validator, indicator.count, false);
-        } else if (indicator.count % misdemeanorThreshold == 0) {
+        } else if (count % misdemeanorThreshold == 0) {
             IBSCValidatorSet(VALIDATOR_CONTRACT_ADDR).misdemeanor(validator);
+        } else if (count >= maintenanceThreshold()) {
```
NathanBSC: count == maintenanceThreshold() ？

otherwise, after validator call exitMaintenance, every slash may cause 'tryEnterMaintenance'

allformless (dismissed):

allformless (approved):

